PR #2, opened 2019-12-12
More time for framework completely stop

alustig3:
Fixes bug that occurred when sending an audio player "stop" message at the end of a session. See: https://groups.google.com/forum/#!topic/pycontrol/sOp8ZIt3lxE

### DIFF
--- a/gui/run_experiment_tab.py
+++ b/gui/run_experiment_tab.py
@@ -222,7 +222,7 @@ class Run_experiment_tab(QtGui.QWidget):
             # Stop running boards.
             if board.framework_running:
                 board.stop_framework()
-                time.sleep(0.01)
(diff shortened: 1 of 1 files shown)


ThomasAkam:
Thanks for the bugfix.  Please make the same modification in _run_experiment_tab.py_ (line 225) then I can merge them both in as a single commit.

alustig3:
> Please make the same modification in run_experiment_tab.py (line 225)

The change I originally made was to line 225. Is there a different line you'd like me to change? line 264 perhaps?

ThomasAkam:
My bad, I had thought that your change was to run_task_tab not run_experiment tab.  Have now changed all time.sleeps after stop framework to 0.05 seconds.

PR #3, opened 2020-01-03
Hide private variables in variables dialog

alustig3:
In some of our tasks we have many variables, including "private" variables that the user neither needs to view nor modify during an experiment. This commit makes it so any variable with a name ending in 3 underscores is considered a "private" variable and will not show up in the variables dialog. This makes the variables dialog less cluttered for more complex tasks.

I've added an example script where we utilized this feature so you can see what I'm talking about. However if this feature were to be merged, It would probably be best to remove that script and maybe modify one of the existing example tasks to include a private variable.

before:
standard

after:
hidden


ThomasAkam:
I like this idea and think it is worth incorporating but am wondering if it should also affect whether variables show up in the GUI when configuring experiments?  This would be more consistent than the current implementation where private variables do not show up in the set variables dialog but do show up when configuring an experiment in the experiements tab.  What do you think based on your use of these?

Also, the current implementation will exclude variables with three underscore anywhere in the name rather than just at the end, it is unlikely that this will make a difference in practice but it is probably best to make the filter as specific as possible, i.e. `if not v_name[-3:] == '___'`

alustig3:
Good catch on the specificity of the underscores. I made the correction.

I agree that the private variables should also be hidden from the experiment tab. I made the changes. 

I removed the markov file that I had earlier as an example and instead updated the reversal_learning.py to have private variables.

ThomasAkam:
I have merged this pull reqest into the dev branch, thanks for the contribution.  In the end I removed private variables from the reversal_learning example because I view them as an advanced feature that may confuse beginners.  I will update the variables section of the programming tasks docs for the next release to detail them.  I plan to do the next release in the next week or so so if you have additional features ready that you want included please submit them in the next few days.

PR #5, opened 2020-01-10
Pause live plot updating

alustig3:
If you want to scroll through the recent history of a plot, or zoom in on a particular sequence of events, it is nice to be able to pause the live updating. With the plot updating paused, the events that you are interested in viewing are no longer moving targets.

This pull request adds a "live updating" checkbox below the plots. It is on by default when you start a session.

### DIFF
--- a/gui/plotting.py
+++ b/gui/plotting.py
@@ -2,7 +2,8 @@ import time
 import numpy as np
 from datetime import timedelta
 import pyqtgraph as pg
-from pyqtgraph.Qt import QtGui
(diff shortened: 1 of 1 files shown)


ThomasAkam:
I think this could be useful but I think it might be more intuitive as a checkable button that pauses the plots.  I implemented this in commit 97fff7c0211d0113749de5199d1efeb7502820f4 pushed to this feature branch.  What do you think of doing it this way?

alustig3:
Your changes look good to me. The button description is definitely an improvement.

PR #7, opened 2020-01-24
Disable save button when there are no changes in the experiment to be saved.

alustig3:


ThomasAkam:
The reason I did not implement this originally is because the logic of enabling/disabling the save button is to provide user feedback on saving, so the behavior is the button is disabled when you load or save an experiment and enabled as soon as you edit it.  The only situation I can see where this change would affect behaviour is if you edit the experiment, and then edit it back to how it was when last saved, and I am not sure it is worth complicating the behaviour of the button for this.

alustig3:
Yes, the edit then undoing edit situation was what prompted me to make the pull request. When an edit is made, the button becomes enabled. To me that indicates that changes have been made and can be saved. If I undo the changes, there is no longer anything new to save, and I think the button should reflect that by becoming disabled again. Also, clicking a the save button when there have been no new changes serves no practical purpose, it just saves an identical copy of what already exists. By disabling the button we don't give the user the option to waste their time. 

Maybe this behavior isn't that intuitive. I think I'm just very used to it working this way in my text editor. When I make changes to a file, a circle indicator appears. 
Screen Shot 2020-01-27 at 11 03 07 PM

If I save or undo those changes, the circle disappears. 
Screen Shot 2020-01-27 at 11 03 38 PM

Anyways, this is a minor detail. Feel free to reject the PR if you disagree with the points above 😄

ThomasAkam:
Have incorporated this in a recent commit.

PR #8, opened 2020-01-27
Experiment flexibility

alustig3:
We train many animals at the same time on the same task (same experiment), but the following requirements for running an experiment were restricting.

1. All subjects are started and run at the same time.
    - We have less experiment rigs than animals, so we can't run them all at the same time even if we wanted to.
2. All subjects are stopped at the same time.
    - Some animals complete their session before others. We'd like the hardware to turn off and the data log to end when the animal is done, not when the last animal of the experiment group is done.

We like the idea of organizing the training sessions by experiment, but we need the ability to start/stop subsets of experimental setups/subjects.

The changes I've made add a "Run" checkbox to the experimental subjects setup. Once the experiment "Run" button is pressed, the selected setups/subjects are initialized. Each setup can then be independently started/stopped. Once the final setup/subject has been stopped, the entire experiment is closed and the experiment setup GUI reappears.

Experiment setup:
setup screen

Experiment loaded:
loaded

Experiment running with individual start/stop
running


ThomasAkam:
I think a lot of this functionallity is useful but it is very important that any changes to how experiments work do not break existing workflows.  A couple of aspects of the current implementation would need to be changed to avoid this:

- The run checkbox for each subject/setup needs to be ticked by default (the default could be controlled by a variable in config/gui_settings so users could change it if desired).

- If only a subset of subjects are being run, I think a dialog should be shown when the run button is pressed to warn the user of this, as it would be very annoying for people to accidentally uncheck a run checkbox and then discover one subject has not run.  This behaviour could also be controlled by a config variable, or alternatively disabled if any setups are repeated in the experiment definition.

-  The start/stop/close button still needs to exist on the run experiments tab but would have slightly different behaviour from previously
    - If no subjects are running the button text reads 'Start all' and the behaviour is to start all subjects running.
    -  If a subset of subjects are running the button is diabled.
    -  If all subjects are running the button text reads 'Stop all' and the behaviour is to stop all subects running.
   - When all subjects have stopped the button text reads 'Close'  and closes the run experiment tab, going back to the configure experiment tab.  I think leaving the run experiments tab open after all subjects have stopped is desirable as this allows the user to inspect the logs, last print statements etc, and it is more natural to show the summary variables dialog over the stopped run experiment tab rather then the configure experiment tab.  

Also, I am quite hesitant to allow a single experiment to have the same setup repeated multiple times with different subjects. I see the use case that you have more subjects than setups, but this can be handled by having multiple experiments that share the same data folder.  If you normally run the same subjects at the same time this is more convenient as you do not have to select the subjets in each cohort manually.  I don't have particularly strong feelings about this though so it you think it is useful I am open to it.  If we do implement this, I want to retain the behaviour whereby when you add a setup/subject it defaults to the next unused setup.

alustig3:
> * The run checkbox for each subject/setup needs to be ticked by default (the default could be controlled by a variable in config/gui_settings so users could change it if desired).

Agreed, changes made.

> * a dialog should be shown when the run button is pressed to warn the user of this, as it would be very annoying for people to accidentally uncheck a run checkbox and then discover one subject has not run. This behaviour could also be controlled by a config variable, or alternatively disabled if any setups are repeated in the experiment definition.

Agreed, changes made.

>* The start/stop/close button still needs to exist on the run experiments tab but would have slightly different behaviour from previously
>   
>   * If no subjects are running the button text reads 'Start all' and the behaviour is to start all subjects running.
>   * If a subset of subjects are running the button is diabled.
>   * If all subjects are running the button text reads 'Stop all' and the behaviour is to stop all subects running.
>   * When all subjects have stopped the button text reads 'Close'  and closes the run experiment tab, going back to the configure experiment tab.  I think leaving the run experiments tab open after all subjects have stopped is desirable as this allows the user to inspect the logs, last print statements etc, and it is more natural to show the summary variables dialog over the stopped run experiment tab rather then the configure experiment tab.

Agreed, changes made.

> Also, I am quite hesitant to allow a single experiment to have the same setup repeated multiple times with different subjects. I see the use case that you have more subjects than setups, but this can be handled by having multiple experiments that share the same data folder. If you normally run the same subjects at the same time this is more convenient as you do not have to select the subjets in each cohort manually. I don't have particularly strong feelings about this though so it you think it is useful I am open to it. 

I think it is useful. Like you said, creating multiple experiments with the subsets of subjects would work, but it requires that you always run the same subjects at the same time, or create many experiments with different subject combinations. Also you would have to maintain multiple experiment variables tables. In theory we do run the same subsets of animals at the same time, but in practice there are situations arise where at least one setup or one subject can't be run (a behavior rig could be malfunctioning, or a rat could be sick). 

>If we do implement this, I want to retain the behaviour whereby when you add a setup/subject it defaults to the next unused setup.

This behavior is retained.

ThomasAkam:
Thanks for making those changes.  I think this functionallity is going to be really useful, but I am inclined to delay integrating it into the main repository till after I release the next version (hopefully in the next few days when I have finished updating the docs).  As this is quite a big change to how experiments work, I want to take the time to try it out for a bit in the lab before releasing it, and do not want to delay the next release any further.

alustig3:
Sounds good!

ThomasAkam:
I have pushed two commits to the experiment-subsets branch.  The main change is to integrate the multithreaded experiment setup developed in the 'multithreaded' branch.  This gives a big speedup for setting up experiments on multiple boxes by doing the setup in parallel rather than serially.  @alustig3 could you take a look at the code and let me know if you are happy with the changes.  If so I will merge it into the dev branch, test it for a couple of weeks in the lab then do annother release.

alustig3:
The changes look good to me.
I tried the multithreaded setup. It works and is a big improvement 👍

alustig3 (commented):
looks good

PR #11, opened 2020-10-13
Task selection menu

alustig3:
Instead of having a combo box, this makes the task selection a dropdown menu. The benefit of this is you can now organize your task files into subfolders within the pyControl task directory. The nested task files can be accessed through submenus. A submenu will only appear if the subfolder within the task directory contains a .py file. An immediate use case would be putting example tasks in an "examples" subfolder. 

I implemented the changes to both the run task tab and the experiment tab.

CleanShot 2020-10-13 at 09 46 04@2x
CleanShot 2020-10-13 at 09 46 21@2x
CleanShot 2020-10-13 at 09 47 08@2x

If a nested item is selected, the text includes the relative path. Should you have tasks with identical filenames but in different subfolders, this will allow you to know which one is selected.

CleanShot 2020-10-13 at 09 46 44@2x


ThomasAkam:
This functionallity is really great, many thanks.   My understanding of the code is that the task menu functionallity is implemented seperately for the configure_experiment_tab and the run_task_tab.  I think it would be cleaner and more maintainable if the task menu code was implemented only once,  in either gui/utility.py or gui/GUI_main.py.  Is there any reason for not doing it this way?

alustig3:
There was no reason for doing it that way, other than I ended up making the changes one tab at a time.  I cleaned up and consolidated the code like you suggested. Should be much better now 😄. Take a look and let me know what you think.

ThomasAkam:
Thanks for refactoring the code.  When I was testing I got errors when trying to run experiments or record data using the run task tab if the task definition file was in a subfolder.    I think I have fixed these issues in the commit I have just pushed.  Could you take a look at the revised code and test it on your machine, and if your happy with it I will merge with the dev branch.

alustig3:
Ah good catch. I was able to replicate the error you described. I tested the revised code and it fixes the problem. Other changes look good to me as well 👍

PR #12, opened 2021-01-23
menu fixes

alustig3:
I came across some odd behavior in the task menu code (#11). 
- The menu was structured incorrectly if there were multiple subfolders
- The menu list was ordered by files, then folders, but was random otherwise. 

These changes fix the subfolder structure, and now the files and folders are presented in alphabetical order.

Here is the actual file/folder structure from my desktop file explorer:
folder_structure

Before, the menu would look like this:
broken

With the changes, it now looks like this:
fixed


ThomasAkam:
Thanks Andy, and apologies for the delay getting to this.

Issue #14, opened 2021-03-01
Display most recent print line in run task tab


lauralgrima:
Hi Thomas, would be great to include the display of the most recent print line (and perhaps the current state and last event) to the run task tab in the way that it's displayed in the run experiment tab. Thanks!

ThomasAkam:
Thanks for the suggestion, I think this is a good idea and will implement it when I get time.

ThomasAkam:
This is now implemented, thanks @alustig3!

PR #22, opened 2021-10-03
Run task recent info display

alustig3:
Added fields to display most recent event, state, and print line. 
Addresses issue #14

run task updates

I also moved the task group to precede the data file group. Since the data group isn't enabled until a task is selected, it makes more sense to me to have the task group come first.


ThomasAkam:
Many thanks for implementing this Andy.  I've added one commit to the branch, the main purpose of which is to factorise out the common code in run_task_tab and run_experiment_tab into a seperate class in utility.py.  Please take a look at the changes and let me know if your happy with them.

alustig3 (commented):
Looks good to me!

PR #23, opened 2021-10-05
Custom variable dialog

alustig3:
## Purpose
This commit gives tasks the ability to have custom dialogs for viewing and adjusting variables. 

## Background
For our behavior tasks we have a large number of parameters that we adjust. The default variable dialog can be unwieldy with so many parameters. Scrolling/searching for parameters, and typing in new values can be tedious or error prone.
CleanShot 2021-10-02 at 14 37 58@2x

To solve this I ended up making a task-specific variable dialog that provides a better GUI for viewing/adjusting variables. With the custom dialog, we can organize parameters into tabbed pages, and adjust them quickly with spinboxes, radio buttons and checkboxes. Video demo below:

https://user-images.githubusercontent.com/8128628/135728701-af73008d-e155-4a4a-8f74-171b8d19235f.mp4

Our fork of pyControl sort of hacks this feature in, but I thought this would be useful for others so I've tried to standardize the process.

## Changes
I did my best to follow the structure established by @neuromantic99's API code https://github.com/neuromantic99/pyControl/tree/api_dev . 

The user places their custom GUI class in the `gui/user_variable_dialogs/` directory, then in their task file assigns their custom class to `v.custom_variable_dialog`

Custom variable dialog GUIs for multiple tasks can be created more easily by importing/reusing elements from a common `dialog_elements.py` file. So far I've only included a standard variable input (acts exactly like the default variable input) a spinbox for bounded integers or floats, and a checkbox for booleans. Some useful features of the dialog elements to point out:
- the spinboxes have a maximum, minimum, and increment setting. 
- you can add suffixes to the spinboxes so values will say "XX ms" or "XX µL"
- you can add "hints" that appear when you hover over the parameter label
- emojis are fair game for the parameter labels, making finding parameters even faster

I have created `custom_variable_gui.py` as a basic example task that uses a custom variable gui.

CleanShot 2021-10-04 at 22 34 11@2x


ThomasAkam:
I can see the utility of being able to customise the variables dialog, and the dialog you have implemented in your main fork is very neat.  However I think the current implementation of a generalised version is likely to be too complex to be attractive to many users.  One way this functionallity could be made very user friendly would be if the user simply created a .csv file detailing how each variable should be displayed (example below based on your blink_gui.py).  This file would then be parsed to generate the custom dialog without the user having to write any code.  This would be quite a bit of work to implement though. What do you think about that approach?

image

alustig3:
I added a slider to the available controls. It has the same parameters as a spinbox.

Filling out the GUI generator like this:
example_generator_table

gives a custom gui that works like this:

https://user-images.githubusercontent.com/8128628/139486123-b90f9913-b619-498e-bd21-5f8abb2b999b.mp4

ThomasAkam:
Thanks very much for the new work on this and apologies for the delay in getting to it.

These custom GUI dialogs have a lot of potential and the GUI generator goes a long way towards making them user friendly enough to be widely used.  I think there are still some functionallity and implementation issues that would be good to address before merging into the dev branch though.

**Implementation:**

- I am not crazy about the GUI generator making and storing the actual code to generate the GUI, because code generation adds an additional level of complexity to the codebase, which will make maintainence and development harder.  I think it would be preferable if the GUI generator interface returned and stored the information needed to create the GUI rather than the code itself - i.e. the contents of the table on the generator dialog, which could be stored this to disk as a csv.  This information would then be passed to a single Custom_variables_GUI class which would contain the code to actually generate the GUI.  This seems a lot simpler than code generation, is there a reason not to implement it this way?

**Critical functionallity:**

- There needs to be a way to modify custom GUIs using the GUI generator dialog after they have been created, otherwise if the user does not get it right first time they have to start again from scratch.  This could be via an edit button in the custom GUI, which would only be available when the task is not running.

- Variables not specified in the GUI generator dialog should have the default representation in the variables dialog, rather than not being included at all.  That way users can add custom controls for the most used variables, but can still adjust other variables if necessary without having to specify them in the generator.  The default ones could be split into a seperate tab to make the custom ones more visable.

**Desirable functionallity:**

- Variable labels should default to the variable name in the GUI builder, perhaps with underscores replaced by spaces.
- Currently the newly created GUI is not used imediately after it has been run but only after the task is reloaded.  Automatically reloading the task after creation / modification of the GUI should fix this.
- Allow users to group variables into different tabs with an optional 'Group'  field in the generator dialog.
- Do we want checkboxes to have a 'get value' button associated with them, or should they be for setting variables only?
- Might be good to have the hints pop up on hover over the control as well as the label.

ThomasAkam:
Thanks Andy.  I can see that enabling users to code their own custom variables dialogs gives the maximum flexibility, and that code generation could make that a bit easier, but overall I think the disadvantages of code generation in terms of complexity and maintainability are not worth it.

> I think a variable GUI generator that can provide sliders, spinboxes, checkboxes and grouped tabs would be super useful to people and would be sufficient for 95% of users, so I will focus on getting that working, using the approach you mentioned of having the generator return/store a json or csv file and then passing that to a Custom_variables_GUI class.

Thanks, I think this will be very useful.

> In the future it could be be nice to allow advanced users to make additional customizations to the variable GUI (pyControl could first look for a .py file with a custom variable GUI class, then fall back to the csv file if not found, and then fall back to opening up the GUI generator dialog if the csv isn't found)

I agree it would be nice to support .py custom variable GUIs to allow maximum customisation for advanced users.  If we provided one example of this, that would probably be enough to show interested users how to get started.

alustig3:
I've made some progress. Tasks below:

**Critical functionality:**

- [x] make custom GUIs editable

- [ ] Variables not specified in the GUI generator dialog should have the default representation in the variables dialog, rather than not being included at all

**Desirable functionality:**

- [x] Variable labels should default to the variable name in the GUI builder, perhaps with underscores replaced by spaces.

- [x] Automatically load/reset the task after creation/modification of the GUI

- [ ] Allow users to group variables into different tabs with an optional 'Group' field in the generator dialog.

- [ ] add 'get value' button to checkboxes, sliders and spinboxes?

- [ ] Have the hints pop up on hover over the control as well as the label.

alustig3:
I think this is pretty close. Would love for others to try it out to expose bugs and give feedback.

There is an included example custom_gui.py task to demonstrate two custom GUIs. Line 30 specifies which example gui is used. 

You can set 
`v.variable_gui = 'blinker_gui' ` 
or
`v.variable_gui = 'blinker_gui_tabs'`
or set it to something different to start from scratch using the GUI editor

ThomasAkam:
Many thanks Andy, this looks great. Sorry for the delay getting to this - I somehow missed the commit and comment when they were made.  I'll take a look ASAP but it may not be before next week.

alustig3:
Hi Thomas,

Thanks for looking over the code. I implemented the changes you suggested:
- [x] move the functions custom_gui_data and open_gui_editor from run_task_tab/run_experiment_tab to custom_variable_GUI
- [x] change some of the names in the code
- [x] Capitalise the first letter of classes (e.g. Checkbox_var not checkbox_var)

> One other question; why use json rather than csv for storing the user created dialogs? csv feels more natural to me given the data is entered in the form of a table, but I don't feel strongly about it.

The real reason I used json is that it was easier in the moment. If I had to make an argument though? There can be multiple variable tabs and thus multiple tables of data and I think it is cleaner/easier to pass around/pack/unpack/handle this data as a nested dictionary rather than a multi-dimensional array. And so if the table data is already a dictionary object, saving dictionaries to json and reading json into dictionaries is super simple. Could have saved as a binary .pkl file instead of json, but filesize is not a concern, and with json, it is still human readable if someone wanted to edit the file without using pyControl GUI. Also, and this is probably a ridiculous edge case to want to account for 😄 ... but I could see a situation where someone goes to edit the csv, it automatically opens in microsoft excel, they make changes, and then it is saved as an .xls file.

If this gets merged, we should remember to add "copy over gui/user_variable_dialogs"  to the [updating pycontrol](https://pycontrol.readthedocs.io/en/latest/#updating-pycontrol) documentation.

ThomasAkam:
This is really awesome, many thanks for implementing it.  I'll merge it into the dev branch now.  I'd like to release this ASAP. Would you be able to write a short section in the GUI docs to make people aware of the functionality?

ThomasAkam:
@alustig3 Looking at the new docs made me wonder if we should move the folder *user_variable_dialogs* to either the config folder or the tasks folder. This would reduce the number of folders that need to be moved when users update the software, and also feels like it would be a more natural place to put it.  I think probably the config folder would be the most approriate location.  What do you think?

alustig3:
I agree that reducing the number of folders to be moved during update is a good idea. Yes config seems appropriate. I will make the changes and create a pull request.

PR #24, opened 2021-10-05
Unify styling across operating systems

alustig3:
## Background
The pyQt UI style for Mac isn't the greatest, it tends to waste a lot of space. Fixes can be made, but it would have side effects for window and linux , or would require adding platform specific code which would be a pain to maintain as the project expands. 

The pyQt UI style for windows is pretty good in my opinion, but still varies a good amount depending on what version of Windows you are running. Also, I feel like things get worse each year with usability being sacrificed for aesthetics most of the time. 

I propose changing to Qt's "Fusion" style so the interface is the same regardless of which OS you are using. Although it doesn't look as sleek as Windows 10, I think the "Fusion" style is quite user friendly. Improvements/customizations will be easy to make since there will only be one style to maintain.

## Changes
- switched to Qt's "Fusion" style and added some custom styling for hovering over buttons and such.
- added the ability to adjust font sizes. User simply edits font variables in `gui_settings.py` 

### Windows before (left) and after (right)
CleanShot 2021-10-04 at 19 58 48@2x

### Mac before (left) and after (right)
CleanShot 2021-10-04 at 20 05 14@2x
le

### DIFF
--- a/config/gui_settings.py
+++ b/config/gui_settings.py
@@ -7,3 +7,7 @@ update_interval = 20 # Interval between calls to the GUIs update function (ms).
 event_history_len  = 200  # Length of event history to plot (# events).
 state_history_len  = 100  # Length of state history to plot (# states). 
 analog_history_dur = 12   # Duration of analog signal history to plot (seconds).
+
+# uncomment either or both of the following variables to customize pyControl's font size
+#ui_font_size = 12 	# the font size of UI elements (button text, label text, line edit text, etc.)
+#log_font_size =  12	# the font size of the pyControl data logger 
--- a/gui/GUI_main.py
+++ b/gui/GUI_main.py
@@ -188,6 +188,28 @@ class GUI_main(QtGui.QMainWindow):
 def launch_GUI():
     '''Launch the pyControl GUI.'''
     app = QtGui.QApplication(sys.argv)
+    app.setStyle('Fusion')
+    style = """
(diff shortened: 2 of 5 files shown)


ThomasAkam:
Thanks Andy.  Unifying the style across platform does seem sensible, particularly as the default style on Mac is not great.  Making the GUI font sizes user configurable is also a good idea.  Two questions though:
- Rather than having the font sizes commented out in the gui config file, is there any reason not to have them uncommented but set to their default values?  This would allow the associated code to be simplified.
- I like the fusion style but am not crazy about the custom blue highlighting, both visually and because it adds more lines of code.  Do you feel strongly about it?

alustig3:
> Rather than having the font sizes commented out in the gui config file, is there any reason not to have them uncommented but set to their default values? This would allow the associated code to be simplified.

At least on my computers, Mac seemed to default to 13, and Windows defaulted to 11, therefore preventing me from making a single default size without deviating from one of the platform's default. But it is much simpler to just choose a default like you suggested, and 11 looks good on both systems, so I made the change.

> I like the fusion style but am not crazy about the custom blue highlighting, both visually and because it adds more lines of code. Do you feel strongly about it?

I don't feel strongly about it. custom highlighting has been removed.

ThomasAkam:
Thanks Andy.  I've modified the default font sizes slightly and will now merge with the dev branch.

PR #25, opened 2021-10-05
changes to run task's  "setup" selection

alustig3:
- "select" label next to a combo box seemed redundant, so I  removed it
- if no setups are found, the combo box now says "No setups found"
- size of combo box now automatically adjusts to width of items, so long setup names won't be cutoff


ThomasAkam:
Looks good, thanks.  When testing it I noticed that the setup names in the dropdown menu were set to be editable (line 56 in run_task_tab).  This was also true of the standard dev branch and does not seem to be desirable.   I have made the board_select dropdown non-editable in the commit I just pushed and will now merge the pull request.

Issue #18, opened 2021-08-03
Errors in hardware timer callbacks fail silently.


ThomasAkam:
If Micropython hardware timer (not a pyControl timer) is set in a pyControl task file (or hardware definition/device driver file?) and the timer callback function generates an error, this error will fail silently without generating an error message or stopping the task from running.

ThomasAkam:
This should be fixed in commit  9858b190837b3303ff6cfcd7aa4b30364ed0492a

PR #30, opened 2021-10-08
Fix to reveal silent ISR failure

alustig3:
I was having a lot of trouble debugging code. I was inheriting the Analog_input class and trying to create a device that reads a distance sensor over I2C. 

My difficulties seemed to be in the same vein as described in this pyControl forum [high frequency events thread](https://groups.google.com/g/pycontrol/c/6MgksnyZCVs). Like @AtMostafa, my task would continue run but the analog plotting was not working and I was getting no error messages.

This commit prints out any bytes that get read, but not processed.
After adding the print statement it ended up revealing this in the terminal running pyControl:

```
uncaught exception in Timer(14) interrupt handler
MemoryError:
```

I was able to fix my interrupt service routine and get everything working.

I think this addresses #18

### DIFF
--- a/com/pycboard.py
+++ b/com/pycboard.py
@@ -395,6 +395,8 @@ class Pycboard(Pyboard):
                     error_message = data_err[:-3].decode()
                     new_data.append(('!', error_message))                
                 break
+            else:
+                print(new_byte.decode("utf-8"),end='')
         if new_data and self.data_logger:
             self.data_logger.process_data(new_data)
         if error_message:


AtMostafa:
Seems like a great idea and should cover many cases. But at some point, I did have an `AttributeError` somewhere and it was being detected as an analog input because of the leading `'A'` in the error message. I guess that's just my bad luck :)

ThomasAkam:
Apolgies for the delay getting to this.  I think it is a good idea to put in place a mechanism for handling unexpected output from the board including these error messages from interrupt handlers which don't actually crash the task.  However I worry that this mechanism is not going to be that robust because if the unexpected output contains the characters 'A', 'D', 'P' or 'V' the pycboard.process_data function is going to interpret them as the start of a valid message and try and process the following characters acordingly.

In commit 9858b190837b3303ff6cfcd7aa4b30364ed0492a I have modified the format of messages sent from the pyboard to the computer during the framework run such that they all start with a special character ( b'\x07'), followed by the 'A', 'D', 'P' or 'V' character indicating the content of the message.  If input not matching this format is recived it is stored in a buffer and then printed in the log as a warning when the next valid message start code is recieved.   This should make data transfer more robust to unexpted output from the board, and I have verified that it catches error messages from ISRs as expected.

Issue #26, opened 2021-10-07
Make analog inputs stream data to computer automatically.


ThomasAkam:
Currently analog inputs have `record()` and `stop()` methods which respectively start and stop streaming the signal to the computer, and they don't stream the signal unless the record method is called from the task file.

In practice it seems very likely that if an analog input is used in a task it is desirable to stream the data to the computer continuously while the task is running.  Making analog inputs automatically stream data to the computer, and removing the functionality to start and stop streaming would simplify both the Analog_input code and task code that uses them.

ThomasAkam:
Now implemented in *new_analog_input* branch.

Issue #16, opened 2021-06-18
'L' data type for subclassed Analog_input causes overflow error in data_logger


ThomasAkam:
The Analog_input class implementation indicates that data type 'L' (unsigned 4 byte integer) is allowed, but this actually causes an overflow in the data_logger class when the data is saved to disk due to conversion to a signed 4 byte integer.

PR #33, opened 2022-02-21
General-purpose data stream

AtMostafa:
I reworked the logic of the `Analog_input` a bit to separate the data streaming from the rest of it. Now, there is a `Data_channel` class that only streams continuous data to the computer and is instantiated inside `Analog_input`.
I made the choice to not inherit from it (I believe this makes more sense) and ideally, event generation should also be independent of the `Analog_input` class (although with changing a few lines, it could be inherited too, if that's preferred).
This PR is not finished yet, in the sense that I am sure that I am breaking some of the API (functions like `_start_acquisition`) and also other hardware modules such as `Rotary_encoder` need to be updated. It is, however, in [working condition](https://github.com/BeNeuroLab/PyTreadmillTask/blob/9c21630f30c508d4a8b392aca8a82b347e23ceae/devices/_PMW3360DM.py#L370) since I managed to inherit the `Analog_input` class and instantiate another `Data_channel` and send 2 independent data streams back to the PC (the _x_ and _y_ coordinates of the mouse sensor).


ThomasAkam:
Thanks @AtMostafa, this looks useful.  I'll take a look ASAP but it may not be before next week.

ThomasAkam:
Apologies for the delay getting to this.  As the changes to the analog input class are likely to be quite extensive (e.g. I'd like to address issues #26 and #16 and possibly also introduce a new analog data format to support things like multidimensional signals), I've created a new branch called new_analog_input to work on them.  Please can you submit this pull request to that branch rather than the master branch.

thanks,

Thomas

ThomasAkam:
Hi @AtMostafa, I have done some work on the new_analog_input branch to make sure that all the functionallity for analog input and rotary encoder (which subclasses analog input) works with the data streaming seperated out.  I also seperated out the functionallity for event generation on threshold crossing as I thought this might come in useful for either implementing multiple thresholds on a signal, or implementing thresholds on multiple components of a >1D signal.  If you have time could you take a look and see if it looks like it would be compatible with your application?  The next step would then be to implement datastreaming functionallity for >1D signals, which would also likely involve reworking the analog data format.

AtMostafa:
Hi @ThomasAkam 
I'll get to it asap, a  bit busy these days so it might take a while.
Cheers,

PR #37, opened 2022-04-12
Fix custom dialog editor "add" button crash

alustig3:
There is a major bug for the custom dialog editor, introduced in c74916061409544a4b618808657bd8b288938e56 . That commit was intended to fix a bug where extra add buttons were getting added underneath the table when an existing custom dialog was being loaded into the custom dialog editor (see video below).

https://user-images.githubusercontent.com/8128628/163068236-cdfcdc7f-f13e-4c6c-8567-1e6db7e93e45.mp4

The hidden buttons problem was fixed, but unfortunately a new bug was introduced. If there are no rows of existing controls loaded into the editor (like when someone is creating a new custom dialog, or adding a new tab to an existing custom dialog), then the add button will add another row but then any further interaction with the editor will cause the entire program to crash (see video below).

https://user-images.githubusercontent.com/8128628/163069620-0e293036-4134-4b64-a015-fe73d722a6c5.mp4

This pull request fixes the crash

### DIFF
--- a/gui/configure_experiment_tab.py
+++ b/gui/configure_experiment_tab.py
@@ -393,7 +393,7 @@ class SubjectsTable(QtGui.QTableWidget):
         remove_button.setIcon(QtGui.QIcon("gui/icons/remove.svg"))
         ind = QtCore.QPersistentModelIndex(self.model().index(self.n_subjects, 2))
         remove_button.clicked.connect(lambda :self.remove_subject(ind.row()))
-        add_button = QtGui.QPushButton('add')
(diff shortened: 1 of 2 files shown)


alustig3:
Hi @ThomasAkam, I'm giving this request a bump since it is a major bug preventing users from creating new custom variable dialogs. On the bright side, I guess no one has complained about it yet in the google group, however it also suggests that no one has tried the new feature haha.

ThomasAkam:
I'll do a release now to fix this bug.

PR #42, opened 2022-09-10
Various GUI updates

alustig3:
### Custom Variable Dialog
- fixed a bug where the slider control could potentially expand to take up too much space
- added a "separator" control that allows you to add horizontal line separators within a control tab.
gui_editor_with_separators
variable_dialog_separator_demo

### Experiment Configuration
- added icons and adjusted layout slightly
config_exp_before
config_exp_after

### Setups
- The "Configure Selected" buttons are now only enabled if there is at least one setup selected

### Other
- fixed typo in log message


ThomasAkam:
Thanks Andy, looks great.

The only change I'm not sure about is the new stretch behaviour of the experiment_select QComboBox.  I think I preferred how it looked when the QComboBox stretched rather than having it fixed width with an empty space beside it that stretches.  Unless you feel strongly about it I'd be inclined to stick with the old behaviour there.

alustig3:
I removed the stretch behavior you referred to.

I also updated the pause plotting button. It now has an icon and the button text updates to show say "Resume plotting" if it is paused, and "Pause plotting" if it is running.

ThomasAkam:
Thanks, I'll merge the pull request now.

PR #46, opened 2022-09-27
Upgrade to PyQt6

alustig3:
I was unable to run pyControl on an M1 mac-mini. There was a lack of support for installing PyQt5 on Apple silicon.

There is support for PyQt6 on Apple silicon, but I needed to make these code changes for pyControl to work. 

Here is a link to some information about the differences between PyQt5 and PyQt6: https://www.pythonguis.com/faq/pyqt5-vs-pyqt6/

The changes should be backwards compatible with PyQt5. I tested on my previous computer that only has PyQt5 and everything worked as far as I could tell. All Qt classes are being imported from PyQtGraph, and PyQtGraph takes care of automatically loading PyQt5, PyQt6, PySide2 or PySide6 from your computer.

### DIFF
--- a/config/user_variable_dialogs/blinker_gui_from_py.py
+++ b/config/user_variable_dialogs/blinker_gui_from_py.py
@@ -5,36 +5,36 @@ from pyqtgraph.Qt import QtGui, QtCore
 from gui.custom_variables_dialog import Slider_var, Spin_var
 
 # Custom Variable dialog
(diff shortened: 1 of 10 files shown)


ThomasAkam:
Thanks Andy.  We'll have to be careful going forward to test against pyqt5 and 6 following changes to the GUI, but I guess it's worth it to maintain cross platform compatibility.

PR #49, opened 2022-10-03
Example tasks import updates

alustig3:
make device driver imports more pythonic as done in 17891ceee0071d1a1c30022bffdae99e91cf2a47

### DIFF
--- a/tasks/example/all_states_example.py
+++ b/tasks/example/all_states_example.py
@@ -5,7 +5,7 @@
 
 # Does not require any hardware except a micropython board.
 
-from pyControl.utility import  goto_state, set_timer, second
+from pyControl.utility import *
 from devices import Digital_output, Digital_input
 
 # Define hardware (normally done in seperate hardware definition file).
--- a/tasks/example/all_states_example.py
+++ b/tasks/example/all_states_example.py
@@ -5,8 +5,8 @@
 
 # Does not require any hardware except a micropython board.
 
-from pyControl.utility import *
-from devices import *
+from pyControl.utility import  goto_state, set_timer, second
+from devices import Digital_output, Digital_input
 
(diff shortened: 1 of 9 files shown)


ThomasAkam:
I'm happy to use named imports for the device classes but am not sure about doing so for pyControl.utility.  I agree it is more pythonic not to use * imports, but either the user has to explicitly import all the names they will be using, or do 'import pyControl.utility as pc' and then access all the pycontrol specific functions and variables as pc.ms etc.  I think either of these options is going to feel less user friendly to people who are not already python programmers.  I think my preference is to stick with * import for pyControl.utility as the default assumption in the docs, but discuss both import options (as is done already) and also have examples that use named imports in the examples folder (we currently have one example but could include a couple more).  What do you think?

alustig3:
I initially made the changes to the device imports to match the what you had done in 17891ceee0071d1a1c30022bffdae99e91cf2a47, thinking when a task file is uploaded, extra devices might be automatically transferred if using `from devices import *`. I should have looked closer/actually tested my concern because I was wrong 😃, the code is smart enough to look at the task file content for the actual instantiation of the class. 

> I think my preference is to stick with * import for pyControl.utility as the default assumption in the docs

I agree the "more pythonic" import is actually less user friendly for the utility import.

I've reverted the changes to the utility imports. I now see the `from devices import *` is not really a problem, so let me know if you'd like me to revert those changes as well. At that point this pull request will just be bugfix for hardware_test.py (https://github.com/pyControl/code/pull/49/commits/f49612f9df875f7267456261c137cf8be76ddcbe)

ThomasAkam:
Thanks for the clarification.  I think I'll keep the devices imports in the examples the same as they were.  As its than only a single file change remaining (hardware_test.py) I'll update it on my machine and push it to the dev branch.

PR #50, opened 2022-10-06
Settings dialog

alustig3:
GUI for adjusting user settings.  Settings are saved to a `user_settings.json` file in the config folder. 
Would love for others to test/review code/give feedback. 

settings_dialog

https://user-images.githubusercontent.com/8128628/194245496-85627599-6b16-4814-94ae-cf10f6c05b2c.mp4


ThomasAkam:
This is great, thanks Andy.  Could you add brief docstrings to the new classes/functions?  Also, I suggest changing the "other" catagory of setting to "GUI" as I think this is more informative.

ThomasAkam:
Many thanks Andy.

Issue #58, opened 2022-10-20
pcx file reasoning


alustig3:
Is there a reason this json data is saved as a .pcx file and not a .json file?

https://github.com/pyControl/code/blob/5d046d8902aba1a96b7ef117eed97d677f8ce364/gui/configure_experiment_tab.py#L218-L221

ThomasAkam:
No - it would make more sense to use a .json extension.  If we change it we should ensure backward compatibility with .pcx extension files.

PR #59, opened 2022-10-21
converted experiment configuration files from .pcx to .json

alustig3:
All reading and writing of experiment configuration files now use .json extension
For backwards compatibility, any .pcx files in the experiments folder are automatically renamed to .json files when pyControl starts up.

This addresses https://github.com/pyControl/code/issues/58


ThomasAkam:
Thanks Andy

PR #60, opened 2022-10-26
Analog input without threshold

alustig3:
Fixes a crash that happens if an analog input is not given a threshold


ThomasAkam:
Good catch, thanks.

PR #64, opened 2022-12-24
digital output pulse updates

alustig3:
This is a first draft of the changes discussed here: https://github.com/pyControl/code/discussions/63

Duty cycle for digital output pulse was previously limited to 10, 25, 50, or 75%
These changes allow the duty cycle to be any multiple of 5 between 5 and 95%

A drawback for this is that the ISR callback can be called up to 20X more frequently than than the pulse frequency. This could potentially be problem if the pulse frequency is high. I haven't yet hooked it up to an oscilloscope to characterize/test the limits. 

What do you think?
- maybe add a warning if the callback frequency is above a certain threshold?
- instead of allowing multiple of five, maybe we make it any multiple of 10 or 25? (basically the performance risks would be the same as before, but we're adding [20,30,40,60,70,80,90] to the existing options of [10,25,50,75] )

### DIFF
--- a/pyControl/hardware.py
+++ b/pyControl/hardware.py
@@ -365,9 +365,6 @@ class Analog_threshold(IO_object):
 # Digital Output --------------------------------------------------------------
 
 class Digital_output(IO_object):
-    freq_multipliers = {10:10, 25:4, 50:2, 75:4}
-    off_inds = {10:1, 25:1, 50:1, 75:3}
-
     def __init__(self, pin, inverted=False):
         if isinstance(pin, IO_expander_pin):
             pin.set_mode(pyb.Pin.OUT)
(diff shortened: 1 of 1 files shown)


ThomasAkam:
Thanks for this and apologies for the delay getting to it.  I pushed a commit to the pulse-improvements branch which fixed a bug where the frequency multiplier was 20 irrespective of the duty cycle.  Also added @micropython.native decorator to the IRS for speed.

I did some testing and it looks like each call of the ISR takes on average 40-70us (depending on the fraction of calls that involve toggling the pin).  For a 50% duty cycle the maximum pulse frequency that can be used without crashing the framework is ~9KHz.  One option would be to warn the user if they specify pulse parameters that will heavily load the framework, e.g. if the ISR frequency is > 2KHz, which should correspond to about 10% of the pyboard processor time being used for the pulsing.  This could be implemented using the 'warning' function which I added in dev branch commit f024b805e36271fd13773bb61f1a5caf52c6cafc.

What do you think?

pyControl/hardware.py
```diff
@@ -410,9 +408,10 @@ def pulse(self, freq, duty_cycle=50, n_pulses=False):
         self.timer.init(freq=freq * self.fm)
```
alustig3, line 408: I agree that a warning here if the timer frequency is > 2KHz seems reasonable.

alustig3 (commented):
Looks good to me

ThomasAkam:
I'll merge this into the dev branch then add the warning.

I also looked into using PWMs at some point but came to the same conclusion that the implementation was going to be too complicated to be worthwhile.

Issue #56, opened 2022-10-14
Enable folders in experiments directory


ThomasAkam:
It would be useful to be able to organise experiments into folders (e.g. by user) similar to what is done for tasks.  Folder organisation in the experiments folder should be reproduced in the data folder when experiment data is saved.

PR #67, opened 2023-01-18
Experiment folders

alustig3:
this commit addresses https://github.com/pyControl/code/issues/56

Creating a new experiment now opens a system create file dialog. Within this dialog you can create new nested folders if you want and name your new experiment.json file. This system file dialog will provide warnings if you are creating a file that already exists. pyControl will throw a warning/not allow you to create an experiment outside of the "pyControl/experiments" folder.

data is organized in the same nested structure as the experiments folder

one behavior that is no longer possible, is that before you could edit the experiment name and save. This would create a copy of the experiment under the new name.

most of the interface is now disabled until an experiment is selected.

Please take a look at the changes and try to break it. I have not had a chance to test on a Windows computer.


ThomasAkam:
Thanks Andy, this looks awesome.

I've done some testing on Windows and the functionality for creating new experiments and accessing them through the nested drop down menu looks like it is working great.  However the functionallity for creating a data folder hierarchy with the same structure does not appear to be working quite right.  If I create an exeriment called `test_folder/test_experiment.json` then the default data directory is `...\data\test_folder` without the `test_experiment` subfolder.  I think this is probably happening because the `new_data_dir` is being created by combining a `Path` object for the data folder with a `str` representation of the `nested_path`, and the `str` is formatted as a Posix path not a Windows path.  I do not know the pathlib module well so am not sure what the best way to fix this is, but I suspect that keeping everyting as `Path` objects rather than `str` might sort it.

A couple of other thought about the implementation:

- It would be useful to have a `Save as` button to allow an existing experiment to be saved with a new name/location.
- Do you think we could put the Data Directory interface on the same row as the Task and Hardware test selection menus?  This would not leave a huge amount of space to display the path, but perhaps we could right justify the path text so the end of it was visible.

PS:  Did you see my post on the discussion board about a new data file format (#66)?  I'd be interested to hear your thoughts about this.

alustig3:
- I think the data folder hierarchy is fixed now. Let me know if it works for you as expected.

- "Save as" functionality has been added

- I rearranged the layout
CleanShot 2023-01-24 at 16 41 30

alustig3:
`get_nested_file_list` docstring has been updated.


I reverted the formatting changes. I will put in a separate pull request for formatting changes in the future.


I am using [Black](https://black.readthedocs.io/en/stable/index.html) formatter. It is an opinionated formatter that gives consistent results and is easy to setup. Here is their [reasoning for double quotes](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings). I don't have strong opinions about it and am open to other formatters, but I do think implementing a formatter of some sort is a good idea for getting consistently clean and readable code.

PR #69, opened 2023-02-02
UI improvements

alustig3:
### Shortcuts
- can now quickly view the error log from the file menu or by using `ctrl + e`
- can open settings dialog with `ctrl + ,` and save changes with `ctrl + s`
- I combined the "Folders" menu items with error log menu item, under a new "View" menu. Maybe I should have left the "Folders" menu untouched and had error log live alone in a separate "View" menu? What do you think?

### Setups tab
- select all / deselect all buttons replaced with select all checkbox
- removed config buttons and replaced with DFU button
     - each setup in the setup table had a config button that would then open a menu with `load framework`, `load hardware definition` etc.
     - this sub menu seems redundant since the same buttons/actions are right below in the `Configure selected` groupbox, and so there are 2 ways to accomplish the same thing, and both approaches require 2 clicks. 
          - approach a: click menu, click load framework
          - approach b: click setup checkbox, click load framework
     - DFU mode is not something you would want to do to multiple setups, so it doesn't make sense to put it with the other multi-select actions.
          - it is now 1 click away, whereas before it was 2
- added a clear log button

### Better large window behavior
The UI was not good when the window was large. It excessively stretched things, hurting readability and made elements spread out and annoying to reach with the mouse. 

I made changes so that for small to medium size windows, the UI is constrained to 1 column width. When the window exceeds 2 column widths, the layout dynamically rearranges to utilize the space better. The column width can be adjusted in the settings dialog (it takes effect after the program is restarted).

https://user-images.githubusercontent.com/8128628/216247426-0a17cb18-d7ea-447a-86a1-0baed3485380.mp4

I hope the changes are seen as improvements. I welcome feedback as always.


ThomasAkam:
Thanks Andy, I like many of these changes but am not sure about some of them:

**Shortcuts:**

-The new shortcuts are good but I think `ctrl+g` might be more memorable for settings than `ctrl+,`, though I don't feel strongly about it.

-The error log viewer is useful.  I'm not crazy about the behaviour of creating an empty error log file if one does not exist though.  The presence/absence of an error log file is a useful indicator of whether an error has occured which is lost by this change.  Also, it might be nice to have a `clear log` button in the dialog.

**Setups tab**

- Select all checkbox is good.
- I can see the logic for removing the configure button but am not crazy about replacing it with the DFU mode button because it really foregrounds this for the user when it is something the user will rarely need to do or even know about.  I think if we are going to get rid of the configure button my preference would be to not replace it with anything but rather to have a DFU mode button below with the other buttons, but disable it if more than one setup is selected.
- I think the row of buttons looks better without the space to the right of it.

**Better large window behavior**

- I think the behavoiur is potentially useful but it comes with a substantial burden of additional complexity to the GUI code which will make maintaining and adding functionallity more complicated and make it less accessible if users want to modify it.  We typically use the GUI on one half of a monitor and find the UI works pretty well in that format.  Is there a particular use case where you find using the GUI in a smaller window is restrictive?  

**Icon changes**
- I'm reluctant to change icons we have been using for a while because once users have got used to one icon then changing it requires them to relearn the icon-function associations.

alustig3:
> The error log viewer is useful. I'm not crazy about the behaviour of creating an empty error log file if one does not exist though. The presence/absence of an error log file is a useful indicator of whether an error has occurred which is lost by this change. Also, it might be nice to have a clear log button in the dialog.

- an empty error log file is no longer created. Instead, an info messagebox will say "You have no errors". 
- I added a "clear log" button that will delete the error log file. A question messagebox will ask for confirmation that you want to clear the error log before proceeding.

 
>  I think if we are going to get rid of the configure button my preference would be to not replace it with anything but rather to have a DFU mode button below with the other buttons, but disable it if more than one setup is selected.

I totally agree and like your solution. Changes made.

>  Is there a particular use case where you find using the GUI in a smaller window is restrictive?

There were two situations I was thinking of where limited vertical real estate might be restrictive, but they are of course not deal breakers, and maybe not worth addressing.
- in the run task tab, If you want to see more of the log at once, your only option right now is to reduce the log font. And if the user has a task with many states and/or many events, as well as an analog plot, the plots can could become too squished and unreadable. 
    - A different potential solution to this would be to add a QSplitter in between, that you could drag to adjust ratio between the log and plot sizes
- in the configure experiment tab, if you have a lot of subjects and a lot of variables this change would again address the limited vertical space and give both tables a lot more space if you were to maximize the window.
    - A different potential solution could be to make the subjects table and variables table into a set of tabs that would take up the majority of the configure experiment window and you could switch between them

Our lab fork already independently has several layout customizations/modifications that we are happy with, so these responsive layout changes won't effect us anyway (i.e. I don't have a strong desire to push these changes if you're on the fence about it). 
CleanShot 2023-02-06 at 15 51 59


 It sounds like the changes wouldn't benefit either of us in particular right now, but maybe other users will chime in in the future if they see limited vertical space as a problem. I've reverted the changes for now.

> I'm reluctant to change icons we have been using for a while because once users have got used to one icon then changing it requires them to relearn the icon-function associations.

Icon changes have been reverted

Thanks for the feedback. Sorry for the flood of changes, I probably should have split these up into separate pull requests.

ThomasAkam:
Thanks Andy, I'll merge this now.  

Regarding the layout, I can see these use cases for where more vertical space would be nice but I'm unsure whether the benefit of the new layout options justifies the additional complication of the GUI code.  I think the options of a QSplitter and tabs are potentially nice solutions.  Other options would be to make it possible to minimize elements (e.g. subjects table or variables table), or enable these to be popped out into seperate windows).  I'm not sure what (if any) of these options would be the best balance of code simplicity and functionallity.

PR #71, opened 2023-02-25
recognize device files that have newer class syntax

alustig3:
Since Python 3, [classes don't need parentheses](https://docs.python.org/3/tutorial/classes.html#class-definition-syntax) if they are not inheriting another class.

This fixes a bug where if you had a device class defined with the new syntax, it would not get recognized and uploaded to the board


ThomasAkam:
Thanks Andy!

PR #75, opened 2023-03-16
Error log improvements

alustig3:
- fixed a bug where the error log dialog would not appear again after you cleared the log, even if more errors later happened. Now instead of deleting the error log file, the clear log button just deletes the contents of the error log file.
- when viewing the error log, the textbox now automatically scrolls to the bottom so you can see the most recent error

### DIFF
--- a/gui/GUI_main.py
+++ b/gui/GUI_main.py
@@ -146,12 +146,20 @@ class GUI_main(QtWidgets.QMainWindow):
         try:
             with open("ErrorLog.txt", "r", encoding="utf-8") as reader:
                 text = reader.read()
-            self.error_log_dialog.log_viewer.setText(text)
-            self.error_log_dialog.exec()
+            if text:
+                self.error_log_dialog.log_viewer.setText(text)
+                self.error_log_dialog.exec()
+            else:
+                QtWidgets.QMessageBox.information(
+                    self,
+                    "No errors",
+                    f"You have no errors",
+                    QtWidgets.QMessageBox.StandardButton.Ok,
+                )
         except FileNotFoundError:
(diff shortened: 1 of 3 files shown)


ThomasAkam:
If possible I would rather not leave an empty error log file after clearing the log as the presence/absence of the error log file in the folder is a useful way to see if an error has occued.

alustig3:
fixed

ThomasAkam:
This does not appear to work on windows, pressing the `Clear log` button in the error log dialog generates an error:

```
  File "C:\pyControl_al3\gui\dialogs.py", line 588, in clear_log
    os.remove(r'ErrorLog.txt')
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'ErrorLog.txt'
```
The docs for `WatchedFileHandler`  say _This handler is not appropriate for use under Windows, because under Windows open log files cannot be moved or renamed - logging opens the files with exclusive locks - and so there is no need for such a handler._

alustig3:
I think the above commit fixes the problem. 

The PermissionError doesn't appear to be specific to using `WatchedFileHandler`. I can get the error on Windows, even if using `FileHandler` and performing the following steps:
- Open pyControl
- force an error, which would then be logged
- view the error log in pyControl and then try to `Clear log`

However if an ErrorLog.txt already exists, and then you open up pyControl and view the errorLog, you are then able to use the `Clear log` button as intended. 

So the act of logging an error once the pyControl program is up and running, ties up the errorLog.txt in the logging fileHandler, and on Windows it won't release it unless `logging.shutdown()` is called.

@ThomasAkam please take a look and let me know if it works.

PR #88, opened 2023-06-09
Hardware setup variables table fixes

### LINKED ISSUE #87: bug in hardware setup variables table
While fixing a bug in 50bd9c11b4d62a2dbf22c14bb9248fb190494929 where running any task or experiment without naming the setups gave an error message, I encountered a bug in the hardware variables dialog.  The bug happens when I open the hardware variables dialog in the setups tab, and causes the error message below to appear in the log:

```
  File "E:\Dropbox\Hardware development\pyControl\gui\hardware_variables_dialog.py", line 113, in refresh_save_button
    if self.get_table_data() != self.starting_table:
  File "E:\Dropbox\Hardware development\pyControl\gui\hardware_variables_dialog.py", line 121, in get_table_data
    setup_name = self.cellWidget(table_row, 0).text()
AttributeError: 'NoneType' object has no attribute 'text'
```

I could not work out what is causing this, @alustig3 could you take a look.

Also,  I have a couple of questions/comments about the code for the `VariablesTable` widget in `hardware_variables_dialog.py`:

- At line [136](https://github.com/pyControl/code/blob/50bd9c11b4d62a2dbf22c14bb9248fb190494929/gui/hardware_variables_dialog.py#L136) what is the purpose of checking `if setup_variables is not False`, and could we refactor it to avoid this somewhat weird syntax?
- At line [130](https://github.com/pyControl/code/blob/50bd9c11b4d62a2dbf22c14bb9248fb190494929/gui/hardware_variables_dialog.py#L130), it is confusing to assign `setup_variables["hw_var"] = self.setup_var_editor.variable_cbox.currentText()` given that all the other entries in the `setup_variables` dictionary are `{setup_name: variable_value}`.  Mixing these different bits of information in the same dictionary then makes the logic of the code in the [save](https://github.com/pyControl/code/blob/50bd9c11b4d62a2dbf22c14bb9248fb190494929/gui/hardware_variables_dialog.py#L134-L151) method hard to follow.  Would it be possible to refactor this, e.g. by just reading the `variable_cbox.currentText()` when the save button is pressed?

alustig3:
Fixes #87

### DIFF
--- a/gui/hardware_variables_dialog.py
+++ b/gui/hardware_variables_dialog.py
@@ -99,7 +99,6 @@ class VariablesTable(QtWidgets.QTableWidget):
         for row, setup in enumerate(self.selected_setups):
             setup_name = QtWidgets.QLabel(setup.name)
             value_edit = QtWidgets.QLineEdit()
-            value_edit.textChanged.connect(self.refresh_save_button)
             if setups_json[setup.port].get("variables"):
                 value = setups_json[setup.port]["variables"].get(hw_variable)
                 if value:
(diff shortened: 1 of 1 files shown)


ThomasAkam:
Thanks Andy!

PR #89, opened 2023-06-15
Experiment code cleanup

alustig3:
- experiment configuration is now a Experiment object instead of a dictionary
- improved some variable and function names


ThomasAkam:
Thanks Andy!

PR #92, opened 2023-07-20
UI layout improvements

alustig3:
@ThomasAkam I made some changes (hopefully seen as improvements) to the UI. No worries if you disagree and don't merge.

## Event trigger
CleanShot 2023-07-20 at 12 15 01

## Experiment configuration tab
- made the subjects and variables tables into tabs to give more vertical space as discussed [here](https://github.com/pyControl/code/pull/69#issuecomment-1419991643)
CleanShot 2023-07-20 at 12 16 56

## Run task tab
- I removed the status because it seemed unnecessary and took up a lot of space. I see how the status can be useful in the experiment view because a lot of information about multiple setups is collapsed/hidden, but in the run task view everything is already in view so the status seems redundant i.e. I never found myself actually looking at the run task status.
- made the top section more compact overall, and user interaction starts in top left and works to the right.

CleanShot 2023-07-20 at 12 12 10

### DIFF
--- a/gui/configure_experiment_tab.py
+++ b/gui/configure_experiment_tab.py
@@ -101,9 +101,10 @@ class Configure_experiment_tab(QtWidgets.QWidget):
         expbox_Vlayout.addWidget(QtWidgets.QLabel("<hr>"))
         expbox_Vlayout.addLayout(self.experiment_parameters_layout)
 
-        # Subjects Groupbox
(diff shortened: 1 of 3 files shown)


ThomasAkam:
Thanks @alustig3, these changes all look very nice, I'll merge them now.

PR #93, opened 2023-07-24
Bug fixes and cleanup

alustig3:
- some formatting and linting fixes
- the shortcuts dialog now shows the appropriate shortcuts for the OS being used
- plotting now fills up layout
- fixed error that would occur if no hardware was connected
- fixed experiment API bug

### DIFF
--- a/com/data_logger.py
+++ b/com/data_logger.py
@@ -63,7 +63,7 @@ class Data_logger:
             os.mkdir(exp_tasks_dir)
         task_file_path = os.path.join(tasks_dir, self.sm_info.name + ".py")
         task_save_name = os.path.split(self.sm_info.name)[1] + "_{}.py".format(self.sm_info.task_hash)
-        if not task_save_name in os.listdir(exp_tasks_dir):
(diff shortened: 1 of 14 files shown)


ThomasAkam:
Thanks @alustig3.  The only change I'm not sure about here is the use of Enums rather than strings for some indicator variables, which seems more complicated than the previous implementation.  What is the rationale for that?

alustig3:
@ThomasAkam the rationale is that it insures at compile-time that the variable/type exists and thus prevents potential mistakes. For instance you could have a conditional statement checking for equality to a misspelled string like having `self.state = "Running"` instead of `self.state = "running"`. In that situation, the code would run without throwing any errors even though the conditional statement isn't what you intended and will always return False. It is also arguably a little cleaner/more readable. 

I don't feel strongly about it and it is probably overkill for this situation considering these variables are only used/accessed a very few times, and within the same file at that.

I've reverted the State(enum) changes.

ThomasAkam:
Thanks @alustig3, that makes sense.  I can see the potential advantage and don't feel stronly either way about using enum here.  Will merge this version for now.

PR #97, opened 2023-09-18
Data format update to "time, type, subtype, content"

alustig3:
I have updated the Datatuple to be time, type, subtype, content. Logged data now contains useful information for several new subtypes. This is particularly important for distinguishing between events coming from task logic, user input from the controls dialog, or code running in an API user_class.

In addition to our discussion [here](https://github.com/pyControl/code/pull/90#issuecomment-1709113225) I have added "publish" and "sync" subtypes. The format is as follows:

| time      | type     | subtype                                      | content                 |
|-----------|----------|----------------------------------------------|-------------------------|
| timestamp | info     | info_type                                    | info_value              |
| timestamp | state    | -                                            | state_name              |
| timestamp | event    | input/timer/publish/user/api/sync            | event_name              |
| timestamp | print    | task/api                                     | print_text              |
| timestamp | variable | get/user_set/api_set/print/run_start/run_end | {"var":val,"var2":val2} |
| -         | warning  | -                                            | warning message text    |
| -         | error    | -                                            | error message text      |

Another notable change is that State and Event types are now separate "S" and "E" types instead of both being "D" type. Accordingly, the Stop type has been changed from "S" to "X"

Below are examples of `example/button.py` data before and after these changes. For each run, I pressed the physical hardware button 3 times, triggered the "button_press" event through the controls dialog 3 times, and then used the controls dialog "get" and "set" a variable.

before:
```
time	type	name	value
0.000	info	experiment_name	run_task
0.000	info	task_name	example/button
0.000	info	task_file_hash	3790482085
0.000	info	setup_id	Rig 1.1
0.000	info	framework_version	2.0rc1
0.000	info	micropython_version	1.2
0.000	info	subject_id	before
0.000	info	start_time	2023-09-18T04:22:54.870
0.000	variable	run_start	{"press_n": 0}
0.000	state	LED_off	
2.087	event	button_press	
2.088	print		Press number 1
2.393	event	button_press	
2.394	print		Press number 2
2.608	event	button_press	
2.609	print		Press number 3
2.609	state	LED_on	
3.609	state	LED_off	
5.702	event	button_press	
5.702	print		Press number 1
5.960	event	button_press	
5.961	print		Press number 2
6.211	event	button_press	
6.212	print		Press number 3
6.212	state	LED_on	
7.212	state	LED_off	
7.951	variable	get	{"press_n": 0}
10.041	variable	set	{"press_n": 2}
11.878	variable	run_end	{"press_n": 2}
11.878	info	end_time	2023-09-18T04:23:06.851
```

after:
```
time	type	subtype	content
0.000	info	experiment_name	run_task
0.000	info	task_name	example/button
0.000	info	task_file_hash	3790482085
0.000	info	setup_id	Rig 1.1
0.000	info	framework_version	2.0rc1
0.000	info	micropython_version	1.2
0.000	info	subject_id	after
0.000	info	start_time	2023-09-18T04:09:00.158
0.000	variable	run_start	{"press_n": 0}
0.000	state		LED_off
2.365	event	input	button_press
2.366	print	task	Press number 1
2.596	event	input	button_press
2.597	print	task	Press number 2
2.869	event	input	button_press
2.870	print	task	Press number 3
2.870	state		LED_on
3.870	state		LED_off
5.305	event	user	button_press
5.306	print	task	Press number 1
5.554	event	user	button_press
5.554	print	task	Press number 2
5.845	event	user	button_press
5.846	print	task	Press number 3
5.846	state		LED_on
6.846	state		LED_off
7.689	variable	get	{"press_n": 0}
11.966	variable	user_set	{"press_n": 1}
14.256	variable	run_end	{"press_n": 1}
14.256	info	end_time	2023-09-18T04:09:14.517
```

`data_import.py` will still need changes before this is ready for release


ThomasAkam:
Hi @alustig3 , many thanks for your work on this and apologies for the delay getting to it.  The changes look great, I have a couple of questions/suggestions:

- In framework.py could we use the type_bytes as the constants used to indicte event types?  So the relevent bit of framework.py would read:
  ```python
  event_typ = b"\x07E"  # External event   : (time, event_typ, [i]nput/[t]imer/[s]ync/[p]ublish/[u]ser/[a]pi, event_ID)
  state_typ = b"\x07S"  # State transition : (time, state_typ, state_ID)
  print_typ = b"\x07P"  # User print       : (time, print_typ, print_string)
  hardw_typ = b"\x07H"  # Harware callback : (time, hardw_typ, hardware_ID)
  varbl_typ = b"\x07V"  # Variable change  : (time, varbl_typ, [g]et/user_[s]et/[a]pi_set/[p]rint/s[t]art/[e]nd, json_str)
  warng_typ = b"\x07!"  # Warning          : (time, warng_typ, print_string)
  stopf_typ = b"\x07X"  # Stop framework   : (time, stopf_type)
  ```
  In the above I have dropped the `timer_typ` as this is now a subtype of event.  We might also want to drop the option of timers not outputting events to the log now that it is easy to filter out timer events from the data.

- What is the `print_msg` method of pycboard used for?  My understanding is that it is used by user APIs to print things to the log, but why does the message need to be sent to the pyboard and then back to the computer?  Is this just to get a pyboard timestamp? 

- I wonder if we should try and standardise the format of both items in the Event_queue and messages passed between pyboard and computer (in both directions) such that event queue items are all tuples of format *(time, type, subtype, content)*, and messages all have a format something like:  *start_byte+type_byte+subtype_byte+content_len+content*.  This should simplify parsing the items and messages.  What do you think?

ThomasAkam:
@alustig3, I've pushed some commits to your subtypes branch which fix a couple of bugs, implement the changes detailed in the first bullet point in my comment above, and modify the framework to use a Datatuple namedtuple for data tuples in event_queues and timers.  Take a look and let me know what you think.

alustig3:
@ThomasAkam, thanks for reviewing and making changes. They look good to me.

> What is the print_msg method of pycboard used for? My understanding is that it is used by user APIs to print things to the log, but why does the message need to be sent to the pyboard and then back to the computer? Is this just to get a pyboard timestamp?

Yes, it is just to get the pyboard timestamp for print messages coming from an API or the controls dialog. I see now that it may be wasteful passing serial data back and forth, especially if the print string is long. Is there a better solution for just grabbing the pyboard timestamp?

> I wonder if we should try and standardise the format of both items in the Event_queue and messages passed between pyboard and computer (in both directions) such that event queue items are all tuples of format (time, type, subtype, content), and messages all have a format something like: start_byte+type_byte+subtype_byte+content_len+content. This should simplify parsing the items and messages. What do you think?

Sorry, I'm not sure I follow. Is this point addressed by the Datatuple you implemented in the framework, or is there something else you would still like to change? I'm not sure how this is different than line 93 , other than it is missing the timestamp and checksum.
`usb_serial.send(b"\x07" + event.type + data_len + timestamp + checksum + data_bytes)`

ThomasAkam:
I've just pushed a commit which substantially simplifies the pycboard.process_data() method by modifying the format of messages sent from the pyboard to allow all messages to be decoded in the same way apart from upacking the 'content' element.  Analog inputs are currently broken as I've not updated their message format yet.

> Yes, it is just to get the pyboard timestamp for print messages coming from an API or the controls dialog

I think it is not ideal to send the whole message to the board and back just for this.  I can see two options for handling it:
1.  We impement a pycboard method to get a timestamp.  The tricky thing with this is that it could not return a timestamp immediately so we would need to then handle the timestamp when it is returned and assign it to the correct print statement.
2. We could compute the time on the computer by recording a datetime whenever a message is recieved from the board along with the pycontrol timestamp, then using the last message datetime and timestamp to compute the current pycontrol time on the computer whenever needed.  
I think option 2 would be my preference.  It does add a small amount of overhead to recieving messages but from a quick test with %timeit it seems to take only about 0.5us to get a datetime.now().

alustig3:
> We could compute the time on the computer by recording a datetime whenever a message is recieved from the board along with the pycontrol timestamp, then using the last message datetime and timestamp to compute the current pycontrol time on the computer whenever needed.

This seems reasonable to me. I made a commit that implements this strategy. Let me know what you think.

ThomasAkam:
Thanks @alustig3 this looks good.   A couple of comments/suggestions:
- It might be preferable to use a `datetime` rather than `time` for the last message time as otherwise there could be an issue if a task was running across a change in date.  Is there a reason to prefer using `time`?
- I suggest using a more descriptive name for the `last_received` variable, e.g. `last_message_datetime`.
- For readability it might make sense to move the functionallity for getting a timestamp currently in `pycboard.print_msg` to a seperate `pycboard.get_timestamp` method.  I don't feel strongly about this though.

Also, do you currently have functionallity implemented somewhere for the user to write a note in the controls dialog which gets printed to the log?  If so it would be great to get this implemented in the default controls dialog.

alustig3:
Thanks @ThomasAkam

> It might be preferable to use a datetime rather than time for the last message time as otherwise there could be an issue if a task was running across a change in date. Is there a reason to prefer using time?

For me `time.time()` was 3X faster than `datetime.datetime.now()`. `time.time()` returns the seconds since jan 1st 1970 UTC, so experiments crossing into a new day is not a problem.

> Also, do you currently have functionallity implemented somewhere for the user to write a note in the controls dialog which gets printed to the log? If so it would be great to get this implemented in the default controls dialog.

yes I have something but I will need to adapt it to make it less specific to our lab and work with the default controls dialog.  I will open a separate pull request for this.

ThomasAkam:
> For me time.time() was 3X faster than datetime.datetime.now(). time.time() returns the seconds since jan 1st 1970 UTC, so experiments crossing into a new day is not a problem.

That makes sense.  I've just pushed a commit reimplementing analog input processing and will merge this into the dev branch now.

ThomasAkam:
@alustig3 I have implemented basic functionallity for adding a note to the log in the dev branch.  Currently it only works when the task is running.

PR #99, opened 2023-10-04
Controls dialog updates

alustig3:
Prevents an error that occurs when pressing the trigger button in the the `blinker.py` example (or any task without events).

- disable events groupbox trigger if no events exist. 
- disable variables groupbox if no variables exist
- move notes groupbox to the top of controls dialog since it will always be enabled


ThomasAkam:
Thanks Andy

PR #102, opened 2023-10-11
Folder renaming and reorganization

alustig3:
This is a reorganization of the repository. Aside from some clearer naming, the major change is towards setting up the user with a separate user folder that contains their api classes, controls dialogs, devices, experiments, hardware definitions, tasks, settings and setups all in one place. A data folder may also be inside the user folder, but there is still an option in the settings to independently change the default save location. 

- A user folder will be nice for lab-specific development. If version controlled, the history will be cleaner and independent of GUI/framework development. 
- Sharing/replicating tasks will be simpler. Just send someone your user folder. Other than the setups.py file which will be specific to your computer's usb ports, everything should work the same on another person's computer.
- Upgrading pyControl will be easier, as you will just need to download the latest release and completely replace the existing pyControl directory. Or git pull the latest, and there shouldn't be any conflicts for a regular user who likely isn't modifying the GUI or framework

Folder structure before:
```
pyControl_code
├── com
├── config
│   ├── user_classes
│   └── user_controls_dialogs
├── data
├── devices
├── experiments
├── gui
│   └── icons
├── hardware_definitions
├── pyControl
├── tasks
│   └── example
├── tests
│   ├── Board tests
│   └── Framework tests
└── tools
```
after:
```
pyControl_code
├── communication
├── gui
│   └── icons
├── pycontrol_mcu
├── tests
│   ├── board_tests
│   └── framework_tests
├── tools
└── user_assets
    ├── api_classes
    ├── controls_dialogs
    ├── devices
    ├── hardware_definitions
    └── tasks
        └── example
```

On startup, the GUI uses [QSettings](https://doc.qt.io/qt-6/qsettings.html) to know where the user folder is. If the user folder has never been setup, or if it has been moved and can't be found, then a dialog will prompt the user to create a new user folder/select an existing one. If the user folder is new (and therefore empty) the device files and example hardware definitions/tasks/api classes/controls dialogs contained in the `user_assets` folder are automatically copied over into the new user folder. A "change user folder" button has been added to the settings dialog.

@ThomasAkam let me know what you think. I realize it is sort of a major shift, and may require significant changes to docs, but I figured now (for v2.0 release that has other breaking changes) would be a good time to do this if we were to consider doing it.


ThomasAkam:
Hi @alustig3. I think that having the ability to have user profiles with their own folders that contain all tasks, experiments, date etc for that user could be a nice feature, and can see it could also simplify aspects of lab-specific development.  However, if we end up doing this I think it should probably be optional to use it.  One way of doing this would be to have a 'User' menu in the menubar, which is used to both select the current user and to add a new user, with a 'default' user which is present when pycontrol is first installed.  When a new user is created the dialog could allow the user to select which user folders are created such that e.g. by default the user has their own 'tasks', 'experiments' and 'data' folder, but they can also optionally have their own 'API', 'control_dialog' and 'hardware_definitions', with these defaulting to the default users if not created (because hardware definitions are unlikely to be user specfic, and the API and custom controls dialog are likely to be used by a minority of power users).  This could result in a folder structure like:

```
pyControl_code
├── communication
├── gui
│   └── icons
├── pycontrol_mcu
├── tests
│   ├── board_tests
│   └── framework_tests
├── tools
└── users
        ├── default
                ├── api_classes
                ├── controls_dialogs
                ├── devices
                ├── hardware_definitions
                └── tasks
                        └── example
```
I think this would work for version controlling lab-specific development in that we could exclude all user folders except 'default' from the main git repo, and individual user folders could be seperately version controlled (though it would not be possible to version control all user folders in a single repo).

What do you think?

alustig3:
Thanks @ThomasAkam for taking a look and providing feedback.

I think the path I was going down, and the corresponding changes needed to address your feedback, was was beginning to get overly complicated. 

I instead reverted some changes and added more path setters to the settings dialog. My desire to have a separately version controlled user folder can still be accomplished this way, while at the same time having the default user experience the same as before. 

CleanShot 2023-10-12 at 18 53 32

As far as having different user profiles, I think the previous "solution" of having nested user subfolders (in the tasks and/or experiments folder) is sufficient.

The folder structure is now:
```
pyControl_code
├── communication
├── gui
│   └── icons
├── pycontrol_mcu
├── tests
│   ├── board_tests
│   └── framework_tests
├── tools
└── user
        ├── api_classes
        ├── controls_dialogs
        ├── data
        ├── devices
        ├── hardware_definitions
        └── tasks
                └── example
```

ThomasAkam:
I think this is a good solution for now.  I would probably be inclined to keep the folders you have put in 'user' at the top level though, to minimise changes for users and keep the folder structure simple.

alustig3:
It's your call, but I think having a separate user folder is a clearer/more organized folder structure.

ThomasAkam:
I think the reasons I am not crazy about this change to the folder structure are:
- It potentially makes it harder for users to find the files the need to work with by moving the relevant directories out of the top level folder.
- The devices folder is not a user folder in the same sense that e.g. the tasks folder is because most users will  be using the default device driver classes and not creating their own.  Similarly the API and Controls dialog folders are power user features that will likely be used by a minority of users.  Putting all these in the user folder may be confusing for new users by creating an impression they need to understand things they can safely ignore for now.

I'm going to ask users in the lab what they think about this as I don't have a good sense for whether it is going to be simpler or more complicated for users in practice.

alustig3:
Another alternative could be the following: 
(`source` my not be the best folder name, but something along those lines)
```
pyControl_code
├── api_classes
├── controls_dialogs
├── data
├── devices
├── hardware_definitions
├── source
│   ├── communication
│   ├── gui
│   │   └── icons
│   ├── pycontrol_mcu
│   └── tests
│       ├── board_tests
│       └── framework_tests
├── tasks
│   └── example
└── tools
```
this would address your concerns while still creating a clear separation between files that a  user could potentially interact with, and files that they should not be concerned with.

my main motivation for creating a separation is to make the upgrade process as easy as possible in the future i.e. "download the latest release and replace the source folder" or "download the latest release and replace the user folder"

Ultimately I think the most user friendly would be `pip install pycontrol_behavior --upgrade` if we can figure that out, so I'm am very interested in the the discussion being had here: https://github.com/pyControl/code/pull/103#issuecomment-1766693686

Maybe the pip package has the source files, and upon startup using QSettings it checks for a user folder and if doesn't exist, it forces the user to choose a location for new or existing local user folder, automatically creating a directory with the requisite folders if needed. I too have never made a pip package so I'm not sure of the best solution. 

It's also hard to tell if the download/upgrade process is even something to be concerned about. It doesn't seem like there has been any issues brought up in the forums/discussions which is good, but I suspect that more has to do with users not trying to upgrade either because they don't know there is a new version available, or they are satisfied with the current version and feel no need to upgrade. Or everybody is on the latest version and had no problem upgrading haha, I have no idea :)

ThomasAkam:
I like this approach of creating a 'source' folder with the non user code and keeping everything else at the top level. @alustig3 please could you make that change and then I'll merge the pull request.

Regarding user behaviour around upgrading to new versions, I suspect that most users don't upgrade often, as in my experience experimentalists are quite reluctant to make changes to working setups. We should nonetheless aim to make the upgrade process easy and trouble free as possible to encourage users to do it. Having pycontrol pip installable might help with this if we can get to work nicely.

ThomasAkam:
Actually, if we are going to go down the pip install route perhaps we should hold off on these changes for now as that is likely to involve a substantial reorganisation.

alustig3:
@ThomasAkam 

I have made an attempt at the python packaging. A test package is currently hosted here: https://test.pypi.org/project/pycontrol-behavior/

you can install on your computer with pip with `pip install -i https://test.pypi.org/simple/ pycontrol-behavior`

once installed, create a new empty folder somewhere that you want to have your user files. then navigate to that folder in your terminal and run the command `pycontrol_init`
This will copy over all of user files into the folder. Your folder should now look like the following:
```
.
├── advanced
│   ├── api_classes
│   ├── controls_dialogs
│   ├── devices
│   └── hardware_definitions
├── data
├── experimenets
├── tasks
├── example
├── tools
└── pyControl_GUI.py
```

You can then launch pyControl by running the pyControl_GUI.py file or from the command line (still at that same directory location) run `pycontrol_run`

## Notes
This is all new to me so I will leave notes here about the process while it is still fresh in my mind

- only .py files are packaged by default, so we need to specify in the toml file to include the .svg icon files

- after `pyproject.toml` file was setup, followed tutorial here: https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives

- prepare system by once by running `python3 -m pip install --upgrade build` and `python3 -m pip install --upgrade twine` 

- build the package with `python -m build`

- you'll need to create a account and get an api key on https://test.pypi.org/account/register/
- upload the package to TestPyPI with `python3 -m twine upload --repository testpypi dist/*`
     - it will ask for username and password. Username should be `__token__` and the password is your api_key

- every time you upload a package to TestPyPI the pyproject.toml version must be unique. Even if you later delete a package version from the server, it won't allow you to replace it, so be careful about uploading versions https://test.pypi.org/help/#file-name-reuse

alustig3:
video of installation

https://github.com/pyControl/code/assets/8128628/9e4700b1-bf8c-4f9f-836a-75b8d960f3d7

@ThomasAkam @MiklosKralik let me know what you think.

ThomasAkam:
@alustig3 thanks for the work on this, I've not had time yet to test it yet but I think this approach seems really promising, I'll try and give it a go and look at he file changes tomorrow.

ThomasAkam:
@alustig3 when I try and pip install using `pip install -i https://test.pypi.org/simple/ pycontrol-behavior` I get the following error:

image

alustig3:
@ThomasAkam 

maybe try pip installing rc1 specifically with the following?
`pip install -i https://test.pypi.org/simple/ pycontrol-behavior==2.0.0rc1`

ThomasAkam:
Hmmm, now I'm getting a different error:

image

alustig3:
is your pip updated to the latest?
`python -m pip install --upgrade pip`

alustig3:
@ThomasAkam did a pip upgrade make a difference? 

When developing this I have been using a Mac, but just now I was able to test the pip installation on a Windows machine and was successful. So at least we know it is not a mac vs. pc problem.

ThomasAkam:
I've not tried on a different machine having upgraded pip and am getting a different error:

image

alustig3:
@ThomasAkam, is the python version >= 3.7? Will it work after you install setuptools?
`pip install --upgrade setuptools`

alustig3:
Hopefully these commands work:

`python -m pip install --upgrade pip setuptools`
`python -m pip install -i https://test.pypi.org/simple/ pycontrol-behavior`

ThomasAkam:
Still getting errors.  I'll try again on annother machine tomorrow.

image

alustig3:
@ThomasAkam, I think I now understand what's going on after looking at this [issue](https://github.com/ultrajson/ultrajson/issues/366)

I was able to replicate your error above when I created a venv and tried the installation process. Basically it is looking for the dependencies (such as numpy) on the _testpypi_ index, but they will not be there. So for testing purposes the installation process should go like this:

`python -m pip install --upgrade pip`
then
`python -m pip install -i https://test.pypi.org/simple/ pycontrol-behavior --extra-index-url https://pypi.org/simple`

After successful installation, navigating to an empty folder and running `pycontrol_init` will create the pyControl user files. Entering `pycontrol_run` from the command line in that same directory will launch pyControl. If the installation was made using the main python pip (and not a venv pip), then you will alternatively be able to launch pyControl by double clicking on the pyControl_GUI.pyw file in the windows file explorer.

once we have the package actually on pypi.org, the installation will be more straightforward
`python -m pip install --upgrade pip`
then
`python -m pip install pycontrol-behavior`

ThomasAkam:
@alustig3 following the instructions in your last message the install worked ok.  Having looked at the implementation I have mixed feelings about whether this is a sensible step to take at this point.

I am not convinced that the installation process is substantially easier using pip.  Installation was already pretty straightforward and the user needs to go through the additional step of making a user folder in addition to the pip install.

It might make upgrading easier but this is not a given as any upgrade that needs to change things in user folders cannot be done by pip alone.  It is not obvious to me that this will be easier or more reliable than using git to pull the latest version of the repository.

It looks like it will make development more complicated,  because the files are now spread across two different locations.  Additionally I think the organisation of the code may be more complicated for new developers to understand.

I don't see advantages to having pycontrol available as a standard python module because it is not usable in other code.  

What are your thoughts about this?

ThomasAkam:
@alustig3 apologies for the delay getting to this.  Good to hear that your in agreement about not moving to pip install at this point, and thanks for your work exploring whether this was a viable option.

I think the addition of the extra folder paths in this pull request is definitely valuable to faciliate lab specific development.  

I'm still unsure about the folder reoranisation for a couple of reasons:

- I'm not sure it will make upgrading simpler, because some upgrades will inevitably require changes to things outside the source folder, precluding having a single instruction (replace 'source' folder) that works in all cases.
- I'm not sure the new folder structure is necessarily simpler for the user to navigate than the previous one, as the only real advantage is the reduction in folders at the top level.  Also, putting some things the user may want to interact with into an 'advanced' folder may reduce the probability of users engaging with them.  Certainly hardware defintions should be at the top level as in my view using them is best practice.

I don't feel particularly strongly about this by my order of preference re folder structure would probably be:
1.  Keep it unchanged.
2. Add a 'source' folder.
3. Add a 'source' and 'advanced' folder.

What do you think?

alustig3:
@ThomasAkam 
I have removed advanced folder. My preference is option 2 with the source folder separating the files and folders that the user shouldn't expect to edit/interact with.

ThomasAkam:
I'm willing to go with option 2.

I'm not keen on changing the user framework imports in task files to `from pyControl_mcu.utility import *`.  My preference would be to stick with `from pyControl.utility` as I don't think it's worthwhile breaking backward compatibilty of task files just to get pep8 compliant on this.

alustig3:
@ThomasAkam I changed it back to `from pyControl.utility`

Just a heads up that I ran into some problems with the pyboard running out of space when testing since at one point there was a `pycontrol` folder as well as a `pyControl_mcu` folder on the pyboard. I found the easiest way to resolve this was to [factory reset the filesystem](https://docs.micropython.org/en/latest/pyboard/tutorial/reset.html#factory-reset-the-filesystem) using the USR and reset button on the pyboard

ThomasAkam:
Thanks Andy, merging now.

PR #103, opened 2023-10-16
Added gui dependencies to .toml Dependencies can now be pip install . from source

MiklosKralik:
Added gui dependencies to pyproject.toml and instructions for pip intallation to readme, this will hopefully start pushing the project in a direction where it is fully pip installable and the files no longer rely on relative imports.


ThomasAkam:
Hi @MiklosKralik, thanks for submitting this pull request.  It could be great to be able to install pycontrol completely using pip, but I have never created a pip installable python package before and am not sure how this would work for a program like pycontrol which in addition to the gui code has various other directories in the folder structure (data, tasks etc) which need to be easily accessible to the user.  Could you explain in a bit more detail about how you envision this working?

Also, why do you want to stop using relative imports, and how would making pyControl PIP installable address this?

MiklosKralik:
Hi @ThomasAkam , making pycontrol pip installable would enable pycontrol to also function as a programmatic interface. The functions would be findable through the $PATH environment variable and could be accessed everywhere. Removing relative imports also makes the repository easier to contribute to as moving files does not create ImportErrors. Making the project pip installable is very simple, you just need to specify module folders and build dependencies in the pyproject.toml file. Once you have this you can register it on the python package index. Special routines, such as micropython installation can be achieved by pointing pip towards specific build scripts. If you would like to develop pycontrol you can install in pip editable mode "pip install -e ."                                  Doing these simple things would not change any functionality of pycontrol, but I think people would be much more willing to contribute to a project that is properly organized.

MiklosKralik:
pyproject.toml configuration:
https://setuptools.pypa.io/en/latest/userguide/quickstart.html

ThomasAkam:
Hi @MiklosKralik, I can certainly see the advantages of making pycontrol pip installable. If you're happy to have a go at this please do go ahead and create a pull request to implement it. Now is a good time to do so, as the next release will be version 2.0, and will make several other substantial changes.

MiklosKralik:
Hi @ThomasAkam This commit makes the repository pip installable from source and automatically installs gui dependencies. It makes the modules accesible from any scripts in the computer. I recommend giving it a quick test. My boards are not currently set up so I cannot do that, but I doubt there are any breaking changes.

MiklosKralik:
The ImportError is probably a quick fix with a try except clause, if the hardware is not available. However I am beginning to see how this project strays away significantly form a classic pip installable python module that it requires a more convoluted implementation. I'll check out some other repositories that seemed to have solved this problem and have both a functional GUI and API.

ThomasAkam:
@MiklosKralik could you take a look at the other open pull request where @alustig3 has done an implementation of making pycontrol pip installable and see what you think.

ThomasAkam:
Closing this, see discussion in #102 regarding pip installation.

PR #107, opened 2023-11-17
Configuration dropdown

alustig3:
## before

CleanShot 2023-11-16 at 22 57 00

## after

CleanShot 2023-11-16 at 22 54 33


ThomasAkam:
Very nice, thanks.

PR #108, opened 2023-11-17
More readable formatting for GUI log

alustig3:
This improves the run start and run end variable display in the GUI log

before:
CleanShot 2023-11-16 at 19 37 02@2x

after:
CleanShot 2023-11-16 at 19 47 56@2x


ThomasAkam:
Thanks @alustig3, this is a nice improvement, but I think would be better implemented in the data logger class rather than in each tab seperately.  This could be done by modifying the [Data_logger.data_to_string](https://github.com/pyControl/code/blob/dev/source/communication/data_logger.py#L105C40-L105C40) method so it has an argument which controls whether the string if for the log or for a data file, allowing for different formatting in the two places.  This would also allow us to format errors better in the log as we would not need to replace  the new lines with `|` as we do currently to avoid breaking the tsv files.

alustig3:
@ThomasAkam the formatting code has been moved to `Data_logger.data_to_string` and now also supports newline formatting for error messages and print messages

ThomasAkam:
Thanks @alustig3 and apologies for the delay in getting to this, the end of year has been hectic but I'll take a look as asap.

T

ThomasAkam:
@alustig3 I've made a couple of small changes to the formatting of variable prints and error messages, could you take a look and let me know what you think.  The changes to error printing were motivated by this behaviour when the GUI window is narrow:

old_error_print

The new behaviour is:

new_error_print

alustig3:
@ThomasAkam your changes look good to me. Thanks!

ThomasAkam:
Great, merging now.

PR #117, opened 2024-01-12
fixed bug where long statements could get lost if followed closely by a variable message

alustig3:
significant bug where some text wouldn't get written to data log. Only noticed once I tried using on a task that had multiple statements printed immediately at startup. 

before I was getting this:
Screenshot 2024-01-12 at 12 00 47 PM

after this fix, I am now getting the full: 
Screenshot 2024-01-12 at 12 00 06 PM

unfortunately this bug was effecting both the GUI log and the saved session.tsv


ThomasAkam:
Well spotted.  Could you submit this pull request directly to the main branch as I want to fix this in a t v2.0.1 release.

ThomasAkam:
Actually, I'll just update the main branch myself.

PR #119, opened 2024-01-29
bugfix, data consumers not processing user and api PRINT messages

alustig3:


ThomasAkam:
Thanks @alustig3 , merging now.

PR #122, opened 2024-06-14
fix disabled hardware variable save button bug

alustig3:
https://github.com/pyControl/code/assets/8128628/1b2c7d03-a6d4-42b8-93b0-86a59bce70f0

If the hardware variables for multiple setups are being edited, the save button will only become enabled if the last variable in the table has changed. This fixes the bug so that the save button becomes enabled if any of the variables is changed.

### DIFF
--- a/source/gui/hardware_variables_dialog.py
+++ b/source/gui/hardware_variables_dialog.py
@@ -106,8 +106,8 @@ class VariablesTable(QtWidgets.QTableWidget):
                     value_edit.setText(str(value))
             self.setCellWidget(row, 0, setup_name)
             self.setCellWidget(row, 1, value_edit)
+            value_edit.textChanged.connect(self.refresh_save_button)
         self.last_saved_dict = self.get_hw_var_dict()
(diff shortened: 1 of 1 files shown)


ThomasAkam:
Hi @alustig3, I somehow missed this and just saw it when looking at the repo for something else. Apologies for the very long delay!

alustig3:
@ThomasAkam, I wanted to mention, for the next time you make a release on GitHub, it might be best to make a fresh release instead of editing the current v2 release.  The only reason I say this is because [this website](https://tooomm.github.io/github-release-stats/?username=pycontrol&repository=code) can track downloads of the releases, which can be useful. But if you edit a release, the download count gets reset and only shows downloads of the latest version of that release (for instance when v2.0.2 was released as an edit to v2.0.1, all of the v.2.0.1 downloads that were counted were lost and the count was set back to 0).

ThomasAkam:
Hi @alustig3, good to know abou that website, I will make sure I only do new releases rather than editing old ones in future.

PR #145, opened 2025-04-03
fix timer conflict between rotary encoder and audio output (again)

alustig3:
A conflict between timers used by the rotary encoder and audio output prevents being used at the same time

This problem was originally found and discussed [here](https://github.com/orgs/pyControl/discussions/70#discussioncomment-5017827)

It was fixed with commit 93a6611206d01b78aad2f7e45c86b3e5b1498f78 in the [v1.8.1 release](https://github.com/pyControl/code/releases/tag/v1.8.1)

Looking through the git history, I am confused/stumped as to where the code regressed. 

From what I can tell the next time `hardware.py` was changed was in ce964c3a9769915b775cfda3b4eb6ce9142051e1, but the regression back to the older list of timers isn't marked as a change...

Likewise, from what I can tell, the next time `audio.py` was changed was in 457c28e83c3bc2ffd0e777ea8561745aba2d9a29, but that commit shows that audio is back to using timer 2, and it is not marked as a change...

Anyway, I reimplemented the same changes as from 93a6611206d01b78aad2f7e45c86b3e5b1498f78

### DIFF
--- a/source/pyControl/audio.py
+++ b/source/pyControl/audio.py
@@ -36,21 +36,21 @@ class Audio_output(hw.IO_object):
     # User functions
 
     def off(self):
-        self._DAC.write_timed(_off_buf, pyb.Timer(2, freq=10000), mode=pyb.DAC.NORMAL)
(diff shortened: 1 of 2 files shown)


ThomasAkam:
Thanks Andy!

PR #148, opened 2025-04-09
Plot height and axis width improvements

alustig3:
## Axis width
Fixed a bug where the tick labels for the plots could potentially not be visible because the axis width wasn't taking into account the GUI font-size

### larger font + long label before:
CleanShot 2025-04-09 at 14 14 03

### after:
CleanShot 2025-04-09 at 14 13 25

## Proportional plot heights
Made the heights of the states and events plots proportional to the number of states and events.

### before:   
CleanShot 2025-04-09 at 14 18 48
### after:
CleanShot 2025-04-09 at 14 19 46


ThomasAkam:
Thanks Andy, this looks great, merging now.

PR #150, opened 2025-05-16
Adds an LED_pulse feature using ISR to `devices/grid_maze.py`

mathias-sm:
Hi everyone,

Thanks for maintaining this great ecosystem!

For an experiment we're running these days, I needed a feature to dim LEDs in a `grid_maze` setup. I'm doing this with PWM using a MicroPython timer and a callback that turns the LED on and off at the required `duty_cycle` (fraction of the time on), at the required `freq`ency. The code is heavily inspired from https://github.com/pyControl/code/blob/a54bdca66567834c78eb6081c0b399a4fa6c0396/source/pyControl/hardware.py#L420-L450 but as they are currently written, this maze's `LED_on` and `LED_off` are allocating new memory with `to_byte` https://github.com/pyControl/code/blob/a54bdca66567834c78eb6081c0b399a4fa6c0396/devices/grid_maze.py#L162-L162 which is incompatible with ISRs (see https://docs.micropython.org/en/latest/reference/isr_rules.html).

Instead, I pre-allocate buffers for on and off, and write those in a tight loop: this works, is as computationally light as I could think of, but makes pulsing incompatible with turning on/off other LEDs or delivering rewards. Feel free to merge, or otherwise leave here should anyone else need this feature! Happy to take the maintainers' input of course.

This is currently being tested in an experiment, so far we have not had any trouble but I will keep upstreaming fixes if need be. Thanks to @ThomasAkam for his input; some of this code was developed together with Arya Bhomick.

Cheers,
Mathias.

### DIFF
--- a/devices/grid_maze.py
+++ b/devices/grid_maze.py
@@ -42,6 +42,8 @@ class Grid_maze(IO_object):
         self.audio = Audio_output(1)
         self.audio.set_volume = self._set_volume
         self.timer = pyb.Timer(hw.available_timers.pop())
+        self.pulse_timer = None  # only ask for a timer if we're going to need one
+        self.stop_pulse = False
 
         # Initialise pins used for pokes that are directly connected to mazehub.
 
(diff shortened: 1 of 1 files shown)


mathias-sm:
Update: it turns out that the race condition between the pulse timer and other `LED_on`/`LED_off`/`SOL_on`/`SOL_off` can, in fact, sometimes occur. Let me fix this before merging.

I now think that the cleanest version of this would be to have a single buffer for the `maze` object, and when we want to communicate with I2C we read _to_ the buffer, update it, and write from it. That way we only ever allocate it once, it can be used in the tight loop, and if the value gets updated outside of the `pulse` callback it's not going to mess up with the rest of the code.

ThomasAkam:
Hi @mathias-sm,

Thanks for your work on this. 

> I now think that the cleanest version of this would be to have a single buffer for the maze object, and when we want to communicate with I2C we read to the buffer, update it, and write from it. That way we only ever allocate it once, it can be used in the tight loop, and if the value gets updated outside of the pulse callback it's not going to mess up with the rest of the code.

If I understand correctly you are proposing the following change to how setting the value of digital outputs works on the gridmaze hardware:
- In the current implementation, when the `write_bit` method is used to change the value of a single bit in a register, the state of the register is first read to a variable using `read_register`, then the bit is toggled in the variable, then the variable is written back to the register using `write_register`.
- You are proposing to maintain the current state of the GPIO registers that control the output pins as an attribute of the Gridmaze class, such that when writing a bit, instead of reading the current register state from the IC, the stored register value is just updated to set the bit and then written to the register.  

I think this is a good plan because it will reduce the overheads associated with writing bits (as it removes the need for an I2C read) and will faciliate controlling outputs in interrupt callbacks.  If you have time to have a go implementing this that would be great.

PR #153, opened 2025-05-28
bugfix note timestamp

alustig3:
If you add a note after a session has started, but before any other events happen, then the note's timestamp will be way off.

below I show what happens if you run `example/button.py` and add a note first thing:
CleanShot 2025-05-28 at 15 10 30

after the fix:
CleanShot 2025-05-28 at 15 18 53


ThomasAkam:
Thanks Andy!

PR #155, opened 2025-05-28
fix missing menu icons on mac

alustig3:
I'm not sure if this bug appeared after a Mac update, or a PyQt update, but I noticed that the menu icons were recently missing on my Mac. I made a small change to fix the problem.

before:
CleanShot 2025-05-28 at 16 34 01
after:
CleanShot 2025-05-28 at 16 34 48


ThomasAkam:
Thanks Andy!

Issue #161, opened 2025-10-29
"Controls" produces a board disconnection


ngarcia-font:
### **Problem:**

In the PyControl GUI, when a task is loaded and you immediately clicks on "Controls", this error appears:

_Error: uncaught exception of type: <class 'AttributeError'>

Disconnected from board._

### **Steps:** 

1. Open the GUI, go to "Run task" and connect to the board
2. Select a task. It can be any task from the 'example' folder that has variables (e.g. button_named_imports).
3. Upload the task.
4. Click on "Controls".
5. A new window with the variables appears and this message is displayed:
 
_Error: uncaught exception of type: <class 'AttributeError'>
Disconnected from board._

### **Comments:**
This error only occurs if the steps are followed as indicated above. If one task is uploaded and run and "Controls" is selected during or after finishing the running, the error does not occur and the controls window opens.

### **Information:**
PyControl board: 
-Micropython version: 1.261
-Framework version: 2.0.2
GUI: pyControl v2.0.2
OS: macOS sequoia version 15.7.1 (also observed in Windows).

alustig3:
Is there anything in your error log that you can share? View->error log or `cmd+e`

ngarcia-font:
Hi Andy. It happens with all the tasks with variables. For example, the button_named_imports which figures in the "example" folder that comes with the package by default, generates the same issue. When I press cmd+e this is what I get:
2025-10-29 17:54:29,547 Traceback (most recent call last):
  File "/Users/ngarcia/Desktop/code/source/gui/dialogs.py", line 35, in showEvent
    bool(board.sm_info.variables) and (board.framework_running or (board.timestamp == 0))
AttributeError: 'Pycboard' object has no attribute 'timestamp'

alustig3:
Great. Could you please attach your `pycboard.py` file. You should be able to drag and drop the file into this github issue comment. the file is under `source\communication\pycboard.py`

ngarcia-font:
[pycboard.py](https://github.com/user-attachments/files/23230213/pycboard.py)

alustig3:
Hi Nuria,

Thank you for finding and reporting this bug!

Please delete your current `pycboard.py` file. Download and replace it with this [pycboard.py](https://github.com/user-attachments/files/23239537/pycboard.py)

Please let me know if that works for you.

ngarcia-font:
Hi Andy,

Thank you very much. I did it and I have the same error:
[...]

PR #163, opened 2025-10-30
controls causing board disconnect

alustig3:
This fixes a bug introduced in https://github.com/pyControl/code/pull/153

The GUI throws and error and the board disconnects if a task (with variables) is uploaded, then before starting the task the controls button is pressed to bring up the controls dialog. 

Thank you @ngarcia-font for reporting the issue https://github.com/pyControl/code/issues/161


alustig3:
@ThomasAkam if you have time, it would probably be good to make a new release

Right now the release and master branch are out of sync. This can get confusing because a user may download a zip of the code from either the https://github.com/pyControl/code `code->download zip` button or the [release page](https://github.com/pyControl/code/releases), making it possible to download different code both claiming to be the same pyControl version.

ThomasAkam:
@alustig3 totally agree a new release is long overdue.  This bug is fixed in the dev branch and there are various other bugfixes and improvements there.  I suggest we merge all changes from dev into the main then do a new release, does that sound sensible to you?

alustig3:
@ThomasAkam yes that sounds good. This bug still exists on the dev branch even after the recent merge with master. The timestamp needs to be initialized during `setup_state_machine()`

ThomasAkam:
I've now fixed this but in the dev branch.  I left the `self.timestamp=0`  in `pycboard.start_framework` to handle situations where the framework is stopped and restarted without resetting the task.

PR #164, opened 2025-10-30
add shortcut to edit selected task  in run task tab

alustig3:
I found this shortcut to be useful. Will open the task file in your default editor


ThomasAkam:
Hi @alustig3,

I think this is useful functionallity except that if the user has the default file association for Python files to be Python.exe, rather than a text editor, the shortcut will fail silently as python tries to run the file and fails.   Some possible options for dealing with this:
i) Merge as is and just add a note to the docs explaining the behaviour.
ii)  Check what program is associated with py files and warn if it is python (not sure if this is possible).
iii) Add a field to the settings allowing the user to specify the path to a text editor to use to open the files.

Thoughts?

PR #166, opened 2025-12-02
simplify UART handler, require Micropython version >=1.26

alustig3:
previously, there was oddness with the pyboard v1.1 (STM32F4) IRQ_RXIDLE
from the [UART docs](https://docs.micropython.org/en/v1.25.0/library/machine.UART.html#machine.UART):

> On STM32F4xx MCU’s, using the trigger UART.IRQ_RXIDLE the handler will be called once after the first character and then after the end of the message, when the line is idle

This difference in how IRQ_RXIDLE worked on pyboard v1.1 and pyboard d-series led to added complexity in UART_handler class where you had to specify `mcu=STM32_F4` or `mcu=STM32_F7`

The extra trigger after the first character has now been fixed in Micropython v1.26
from [release notes](https://github.com/micropython/micropython/releases/tag/v1.26.0):
> extmod_hardware/machine_uart_irq_rxidle.py: ignore inital IRQ

This commit has changes that simplify the UART_handler class, with the caveat that it now requires Micropython >= 1.26

### DIFF
--- a/devices/uart_handler.py
+++ b/devices/uart_handler.py
@@ -1,3 +1,4 @@
+import sys
 from pyControl.hardware import IO_object, assign_ID, interrupt_queue, fw, sm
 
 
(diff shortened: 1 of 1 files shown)


ThomasAkam:
Thanks Andy

PR #178, opened 2026-07-28
Add pyproject.toml for project configuration

juan-cobos:
Add the dependencies suggested in the docs to a pyproject.toml compatible with uv package manager for easier installation.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ name = "code"
 version = "0.1.0"
 description = "Add your description here"
 readme = "README.md"
-requires-python = ">=3.13"
+requires-python = ">=3.11"
 dependencies = [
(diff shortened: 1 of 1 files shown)


ThomasAkam:
Hi @juan-cobos,

Thanks for suggesting this, it sounds like a good idea.  What is the rationale for putting the requirements in pyproject.toml rather than in a requirements.txt?

Also, do we need to specify version numbers, currently the [installation](https://pycontrol.readthedocs.io/en/latest/#dependencies) docs do not do.

@alustig3  - any thoughts about this, I don't have experience with uv.

juan-cobos:
@ThomasAkam It's a convention so tools like uv find them automatically. Also the version is to ease the dependency resolution with other packages.

